PR #2, opened 2022-06-16
DY-1 transformation engine infrastructure

spicy-sauce:


src/datayoga/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Returns:
+        Job: Compiled job
+    """
+    logger.debug("Compiling job")
+    job = Job(job_yaml)
+    return job
+
+
+def transform(job: Job, data: Any, context: Context = None) -> Any:
```
zalmane: I think we can remove this.
if you use dy.compile you get a Job and can then call its 'transform' method.
If we want 'transform' as dy.transform, it should receive the Yaml as a convenience to skip the 'compile' part separately.

src/datayoga/block.py
```diff
@@ -0,0 +1,79 @@
...
+        self.properties = properties
+        self.validate()
+        self.init()
+
+    def validate(self):
+        """
+        Validates block against its JSON Schema
+        """
+        json_schema_file = path.join(os.path.dirname(os.path.realpath(
+            sys.modules[self.__module__].__file__)), "block.schema.json")
```
zalmane, line 39: does this work after we packaged it as a wheel?

src/datayoga/block.py
```diff
@@ -0,0 +1,79 @@
...
+        Args:
+            data (Any): Data
+
+        Returns:
+            Any: Transformed data
+        """
+        pass
+
+    def get_block_name(self):
+        return self.__class__.__name__
```
zalmane, line 79: how does this get set? what is it used for?

src/datayoga/utils.py
```diff
@@ -0,0 +1,52 @@
+import json
+import re
+from typing import Any, Dict
+
+import yaml
+
+
+def snake_to_camel(value: str) -> str:
```
zalmane: where do we use camelCase?

src/datayoga/__init__.py
spicy-sauce: done

src/datayoga/block.py
spicy-sauce: used in the block implementation, for example:
```logger.debug(f"Initializing {self.get_block_name()}")```

File: src/datayoga/utils.py (hunk 4 above)
Comment (spicy-sauce) [reply src/datayoga/utils.py]: In the yaml file we use `remove_field` while the class itself named `RemoveField`

File: src/datayoga/block.py (hunk 2 above)
Comment (spicy-sauce) [reply src/datayoga/block.py:39]: yes

File: src/datayoga/__init__.py [hunk 5]
```diff
@@ -22,17 +22,18 @@ def compile(job_yaml: Dict[str, Any]) -> Job:
...
     Args:
-        job (Job): Job to transform against
+        job_yaml (Dict[str, Any]): Job in YAML format
         data (Any): Data to transform
         context (Context, optional): Context. Defaults to None.
 
     Returns:
         Any: Transformed data
     """
+    job = compile(job_yaml)
```
zalmane: this should not be "yaml" if we are not reading the source itself. it is just a dict, so Json and YAML are the same here

spicy-sauce: right, fixed

PR #6, opened 2022-06-20
DY-5 map block

### LINKED ISSUE #5: `map` block

spicy-sauce:


src/blocks/add_field/block.py
```diff
@@ -1,43 +1,31 @@
...
     def init(self):
         logger.debug(f"Initializing {self.get_block_name()}")
-        self.language = self.properties.get("language", Language.JMESPATH.value)
+        self.language = self.properties["language"]
+
         if self.language == Language.JMESPATH.value:
             self.expression = jmespath.compile(self.properties["expression"])
         elif self.language == Language.SQL.value:
-            self.conn = sqlite3.connect(":memory")
+            self.conn = get_connection()
```
zalmane, line 21: this should actually remain hardcoded since it is not really creating a connection but rather using SQLite for expression parsing.
Perhaps the utils method should be to compile/execute an SQL expression similar to JMESPath.

src/blocks/add_field/block.py
```diff
@@ -1,43 +1,31 @@
...
         elif self.language == Language.SQL.value:
-            clauses = []
-            for k, v in data.items():
-                clauses.append(f"'{v}' as '{k}'")
-
-            from_clause = f"select {','.join(clauses)}"
-
-            value = self.conn.execute(f"select {self.properties['expression']} from ({from_clause})").fetchone()[0]
-            data[self.properties["field"]] = value
+            data[self.properties["field"]] = exec_sql(self.conn, data.items(), self.properties["expression"])
```
zalmane, line 29: See above. I would move the connection into the exec_sql

zalmane: and maybe like JMESPath we have a 'compile' that essentially holds the :memory connection for better performance

src/datayoga/blocks/utils.py
```diff
@@ -0,0 +1,32 @@
...
+def get_connection() -> Connection:
+    """Gets sqlite3 in memory connection
+
+    Returns:
+        Connection: sqlite3 connection
+    """
+    return sqlite3.connect(":memory")
+
+
+def exec_sql(conn: Connection, fields: Any, expression: str) -> Any:
```
zalmane, line 15: can you add a test for utils.exec_sql? even a basic one

PR #10, opened 2022-06-21
8 unit test for exec sql

### LINKED ISSUE #8: unit test for exec_sql

spicy-sauce:


tests/test_utils.py
```diff
@@ -0,0 +1,8 @@
+from datayoga.blocks import utils
+
+
+def test_exec_sql():
+    assert utils.exec_sql(
+        utils.get_connection(),
```
zalmane, line 6: why do we need the utils.get_connection()?
the in memory connection is just an implementation detail of the exec_sql used as an expression language. We could replace ths in the future with just parsing the SQL text. I would embed it in the exec_sql, and split it into a .compile and run, like we have with JMESPath

spicy-sauce: will be done as part of #9

PR #118, opened 2022-10-09
100 e2e integration tests | 111 read_csv should search for data in the data folder | 105 generate_docs action: run jsonschema2mk on resources/schemas directory

spicy-sauce:


datayoga-py/datayoga/blocks/read_csv/block.py
```diff
@@ -12,8 +13,14 @@ class Block(DyBlock):
 
     def init(self, context: Optional[Context] = None):
         logger.debug(f"Initializing {self.get_block_name()}")
+        csv_file = self.properties["file"]
 
-        self.file = self.properties["file"]
+        if os.path.isabs(csv_file) or context is None:
+            self.file = csv_file
+        else:
+            self.file = os.path.join(context.properties.get("job_path"), csv_file)
```
zalmane: this should not be the job path. It should actually be a 'data path' that comes in from the config and has some value that can be overrided somehow if needed.

zalmane: probably a property of the runner

datayoga-py/datayoga/blocks/read_redis_stream/block.py
```diff
@@ -21,6 +21,7 @@ def init(self, context: Optional[Context] = None):
             connection.get("password"))
 
         self.stream = self.properties["stream_name"]
+        self.read_once = self.properties.get("read_once", False)
```
zalmane: what does this property mean? saw it before and didn't understand

spicy-sauce: This is used in the integration test, an indication to read from the stream only once and exit instead of the infinite loop.

datayoga-py/datayoga/blocks/read_csv/block.py
spicy-sauce: data_path is better than job_path, I agree, will change.

zalmane:
where are these defined?

spicy-sauce:
> where are these defined?

In the run command (__main__.py):
```
  context = Context({
            "connections": connections,
            "job_name": Path(job_file).stem,
            "job_path": job_path
        })
```

datayoga-py/datayoga/blocks/read_csv/block.py
spicy-sauce: done

zalmane (approved):

PR #152, opened 2022-11-03
`std.read` block#144 | `std.write` block#145

spicy-sauce:


datayoga-py/datayoga/blocks/std/read/block.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class Block(DyProducer):
+
+    def init(self, context: Optional[Context] = None):
+        logger.debug(f"Initializing {self.get_block_name()}")
+
+    async def produce(self) -> Generator[Message, None, None]:
+        def get_records(data: str) -> List[Dict[str, Any]]:
+            records = ast.literal_eval(data)
```
zalmane: what is this for? the literal_eval always seems freightnin because of injection risk (although we went through it before and it is probably safe). I would use something explicit. Or remove completely.

datayoga-py/datayoga/blocks/std/read/block.py
```diff
@@ -0,0 +1,42 @@
...
+
+            if isinstance(records, dict):
+                records = [records]
+
+            return records
+
+        def issue_message(record: Dict[str, Any]) -> Message:
```
zalmane: what does this method do?

datayoga-py/datayoga/blocks/std/read/block.py
```diff
@@ -0,0 +1,42 @@
...
+logger = logging.getLogger("dy")
+
+
+class Block(DyProducer):
+
+    def init(self, context: Optional[Context] = None):
+        logger.debug(f"Initializing {self.get_block_name()}")
+
+    async def produce(self) -> Generator[Message, None, None]:
+        def get_records(data: str) -> List[Dict[str, Any]]:
```
zalmane: does this have to be nested within the method vs a class method?

datayoga-py/datayoga/blocks/std/read/block.py
```diff
@@ -0,0 +1,42 @@
...
+
+            if isinstance(records, dict):
+                records = [records]
+
+            return records
+
+        def issue_message(record: Dict[str, Any]) -> Message:
+            return {self.MSG_ID_FIELD: str(uuid.uuid4()), **record}
+
+        if select.select([sys.stdin, ], [], [], 0.0)[0]:
```
zalmane: please add comment. I assume this checks if someone is piping data vs interactive?

datayoga-py/datayoga/blocks/std/write/block.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Block(DyBlock):
+
+    def init(self, context: Optional[Context] = None):
+        logger.debug(f"Initializing {self.get_block_name()}")
+
+    async def run(self, data: List[Dict[str, Any]]) -> Tuple[List[Dict[str, Any]], List[Result]]:
+        for record in data:
+            sys.stdout.write(f"{record}\n")
```
zalmane: do we have an easy way to print a nice table as an alternate format?

datayoga-py/datayoga/producer.py
```diff
@@ -1,18 +1,19 @@
...
 from .block import Block
 
 logger = logging.getLogger("dy")
+Message = TypedDict("Message", msg_id=str, value=Dict[str, Any])
 
 
 class Producer(Block):
 
-    def produce(self) -> List[TypedDict("Message", msg_id=str, value=Dict[str, Any])]:
+    async def produce(self) -> Generator[Message, None, None]:
```
zalmane: why does the produce need to be async?

datayoga-py/datayoga/tests/test_transformation_engine.py
```diff
@@ -55,8 +55,8 @@ def test_validate_valid_job(job_settings):
 
 
 def test_validate_invalid_job():
-    # `with` key is missing in this block
-    job_settings = {"steps": [{"uses": "add_field"}]}
+    # unsupported property specified this block
```
zalmane: specified -> "specified in"

spicy-sauce: done

datayoga-py/datayoga/producer.py
spicy-sauce: because of this in the std.read producer:
```
data = await aioconsole.ainput()
```

datayoga-py/datayoga/blocks/std/write/block.py
spicy-sauce: why? you would probably like to pipe it to a file and have a newline delimited json as you suggested...

datayoga-py/datayoga/blocks/std/read/block.py
spicy-sauce: exactly, done.

datayoga-py/datayoga/blocks/std/read/block.py
spicy-sauce: no, changed.

datayoga-py/datayoga/blocks/std/read/block.py
spicy-sauce: returns a dict with the message id and the payload. I've renamed it to `get_message` instead.

datayoga-py/datayoga/blocks/std/read/block.py
spicy-sauce: it takes a string and converts it into a dict without caring to much if ' or " is used (while json.loads shouts on one of them). What do you suggest to use instead?

zalmane: fix to use json.loads

datayoga-py/datayoga/producer.py
zalmane: let's find out why producer needed to be async

datayoga-py/datayoga/blocks/std/read/block.py
spicy-sauce: fixed

datayoga-py/datayoga/blocks/std/write/block.py
spicy-sauce: will consider this as an option (will report another issue)

datayoga-py/datayoga/producer.py
spicy-sauce: reverted

zalmane (approved):

PR #167, opened 2022-11-15
add support for nested dicts in SQL

zalmane:


spicy-sauce:
check why the tests fail...

spicy-sauce (approved):
@zalmane see my last commit, if it's OK, we can merge.

PR #192, opened 2022-12-05
`relational.write` block: operation to be determined based on `opcode_field` #191

### LINKED ISSUE #191: `relational.write` block: operation to be determined based on $$opcode field
If specified, assume that this field exists in the data and contains:
c - create
d- delete
u - update

It also means that this determines the load strategy, so the load strategy is not requested if the `opcode_field` is passed.

spicy-sauce:
upsert support for postgresql

### DIFF
--- a/core/src/datayoga_core/blocks/relational/write/block.py
+++ b/core/src/datayoga_core/blocks/relational/write/block.py
@@ -94,11 +94,13 @@ class Block(DyBlock):
                 data, key=lambda record: record.get(self.opcode_field, "").replace(OpCode.CREATE.value, OpCode.UPDATE.value))]
 
             for records_by_opcode in opcodes:
+                opcode = records_by_opcode["opcode"]
                 records = records_by_opcode["records"]
-                if records_by_opcode["opcode"] == OpCode.UPDATE.value:
+
+                if opcode == OpCode.UPDATE.value:
                     logger.info(f"Upserting {len(records)} record(s) to {self.table} table")
                     self.conn.execute(self.upsert_stmt, records)
-                else:
+                elif opcode == OpCode.DELETE.value:
                     logger.info(f"Deleting {len(records)} record(s) from {self.table} table")
 
                     keys_to_delete = []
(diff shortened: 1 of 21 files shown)


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -10,14 +12,48 @@
...
+        db_type: str) -> Any:
+    if db_type.lower() == "postgresql":
+        update_fields = [f"{field['column']} = {sa.bindparam(field['key'])}" for field in mapping_fields]
+
+        insert_fields = ", ".join([field["column"] for field in mapping_fields])
+        pk_fields = ", ".join([field["column"] for field in primary_keys])
+
+        insert_bind_params = ", ".join([f"{sa.bindparam(field['key'])}" for field in mapping_fields])
+
+        return sa.text(f"""INSERT INTO {table} ({insert_fields})
```
zalmane, line 40: there is an sql builder for the upsert now in SA. https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#insert-on-conflict-upsert.
we should move to the API in next iteration as we add more dialects.

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -27,17 +63,67 @@ def init(self, context: Optional[Context] = None):
...
+            # group records by opcode
+            opcodes = [{"opcode": key, "records": list(result)} for key, result in groupby(
+                data, key=lambda record: record.get(self.opcode_field, "").replace(OpCode.CREATE.value, OpCode.UPDATE.value))]
+
+            for records_by_opcode in opcodes:
+                opcode = records_by_opcode["opcode"]
+                records = records_by_opcode["records"]
+
+                if opcode == OpCode.UPDATE.value:
+                    logger.info(f"Upserting {len(records)} record(s) to {self.table} table")
```
zalmane: change to debug

spicy-sauce: done

core/src/datayoga_core/blocks/relational/write/block.py
spicy-sauce: I'll open an issue for that.

zalmane (approved):

PR #195, opened 2022-12-06
#174 `in` custom function for jmespath

### LINKED ISSUE #174: `in` custom function for JMESpath
Example:
in(country, ['IL', 'US', 'RU'])
--> True/False

vd2org:


core/README.md
```diff
@@ -125,18 +125,19 @@ DataYoga supports both SQL and [JMESPath](https://jmespath.org/) expressions. JM
...
+| `hash`               | Calculates a hash using the `hash_name` hash function and returns its hexadecimal representation                                     | Input: `{"some_str": "some_value"}` <br /> Expression: `hash(some_str, &#96;sha1&#96;)` <br /> Output: `8c818171573b03feeae08b0b4ffeb6999e3afc05` | Supported algorithms: sha1 (default), sha256, md5, sha384, sha3_384, blake2b, sha512, sha3_224, sha224, sha3_256, sha3_512, blake2s                              |
+| `in`                 | Checks if the iterable contains the given element                                                                                    | Input: `{"el": "b"}` <br /> Expression: ``in(el, `["a", "b", "c"]`)`` <br /> Output: `True`                                                       |                                                                                                                                                                  |
```
spicy-sauce: Change description to:
Checks if an element matches any value in a list of values

core/src/datayoga_core/expression.py
```diff
@@ -143,10 +143,10 @@ def compile(self, expression: str):
         self.expression = jmespath.compile(expression)
         self.filter_expression = jmespath.compile(f"[?{expression}]")
 
-    def filter(self, data: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    def filter(self, data: Union[Dict[str, Any], List[Dict[str, Any]]]) -> Any:
```
spicy-sauce, line 146: Shouldn't it return:
Optional[List[Dict[str, Any]]
?

core/src/datayoga_core/expression.py
```diff
@@ -143,10 +143,10 @@ def compile(self, expression: str):
...
         return self.filter_expression.search(data, options=self.options)
 
-    def search(self, data: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    def search(self, data: Union[Dict[str, Any], List[Dict[str, Any]]]) -> Any:
```
spicy-sauce, line 149: Shouldn't it return:
Optional[List[Dict[str, Any]]
?

spicy-sauce (changes requested):

core/src/datayoga_core/expression.py
vd2org: It could return anything depends on call. In tests it returns direct value, for example.

core/README.md
vd2org: ✅

spicy-sauce (approved):

PR #201, opened 2022-12-12
reason per record, not per batch #147

### LINKED ISSUE #147: reason per record, not per batch
```
    def done(self, msg_ids: List[str], result: Optional[Result] = None, reason: Optional[str] = None):
```

spicy-sauce:


core/src/datayoga_core/block.py
```diff
@@ -4,21 +4,19 @@
 import logging
 import os
 import sys
-from enum import Enum
+from abc import ABCMeta, abstractmethod
```
zalmane, line 7: what is this ABCMeta?

core/src/datayoga_core/blocks/add_field/tests/test_add_field.py
```diff
@@ -21,7 +22,7 @@ async def test_add_nested_field():
                   "expression": "[name.fname,name.lname] | join(' ', @)"})
     block.init()
     assert await block.run([{"name": {"fname": "john", "lname": "doe"}}]) == ([
-        {"name": {"fname": "john", "lname": "doe", "full_name": "john doe"}}], [Result.SUCCESS])
```
zalmane, line 24: why the change from the Enum?

core/src/datayoga_core/blocks/add_field/tests/test_add_field.py
```diff
@@ -1,7 +1,8 @@
 import pytest
-from datayoga_core.block import Result
 from datayoga_core.blocks.add_field.block import Block
+from datayoga_core.result import Result
 
+RESULT_SUCCESS = Result.success()
```
zalmane: why a method and not a plain enum?

core/src/datayoga_core/result.py
```diff
@@ -0,0 +1,26 @@
...
+        self.message = message
+
+    def __eq__(self, other):
+        return (
+            isinstance(other, self.__class__) and
+            self.status == other.status and
+            self.message == other.message)
+
+    @staticmethod
+    def success() -> Result:
```
zalmane: why a static method and not a contstant/enum?

core/src/datayoga_core/utils.py
```diff
@@ -88,3 +90,17 @@ def get_connection_details(connection_name: str, context: Context) -> Dict[str,
             return connection
 
     raise ValueError(f"{connection_name} connection not found")
+
+
+def produce_data_and_results(data: List[Dict[str, Any]]) -> Tuple[List[Dict[str, Any]], List[Result]]:
```
zalmane, line 96: what is the purpose of this method?

integration-tests/test_stdin_to_stdout.py
```diff
@@ -4,7 +4,7 @@
 from common.utils import run_job
 
 
-def test_stdin_to_stdout(tmpdir: str):
```
zalmane, line 7: any reason this was changed?

core/src/datayoga_core/block.py
spicy-sauce: https://docs.python.org/3/library/abc.html

This module provides the infrastructure for defining [abstract base classes](https://docs.python.org/3/glossary.html#term-abstract-base-class) (ABCs) in Python

core/src/datayoga_core/blocks/add_field/tests/test_add_field.py
spicy-sauce: It's not Enum anymore, it's a Result object with SUCCESS status.

core/src/datayoga_core/blocks/add_field/tests/test_add_field.py
spicy-sauce: answered in a different comment

core/src/datayoga_core/result.py
spicy-sauce: answered in a different comment

core/src/datayoga_core/utils.py
spicy-sauce: it goes over the records and puts the internal result field (Block.RESULT_FIELD) in the results array. If it's not specified, it's success. Then, it removes this field from the record.

integration-tests/test_stdin_to_stdout.py
spicy-sauce: yes, it's not a str, but a pytest object and it was confusing to see read() on str.

core/src/datayoga_core/blocks/add_field/tests/test_add_field.py
zalmane: Maybe use @dataclass unless we actually have business logic in those? This mode of a 'factory' singleton just seems unnatural. especially since we are creating one as a constant, but then duplicating it.

spicy-sauce: From @zalmane :
Here is my suggestion. I would not mix static and non static methods in that dataclass. It is really meant as a Struct or NamedDict:
result.py
success= Result(Status.SUCCESS)
@dataclass
class Result():
    status: Status
    message: Optional[str] = None
  
    @staticmethod - remove this altogether. the code can just do this one liner and makes it more readable and easier to maintain
    def reject(message: Optional[str] = None) -> Result:
        return Result(Status.REJECTED, message)
so you use it:
import result
result.success

spicy-sauce: done

zalmane (approved):

PR #212, opened 2022-12-16
`relational.write`: changes in upsert and delete statements  #211

### LINKED ISSUE #211: UPSERT - only fields listed in `mapping` should get updated

spicy-sauce:
- UPSERT - only fields listed in mapping should get updated 
- business keys are now added to mapping
- do not use `sa.text` on delete

### DIFF
--- a/core/src/datayoga_core/blocks/relational/write/block.py
+++ b/core/src/datayoga_core/blocks/relational/write/block.py
@@ -34,19 +34,12 @@ def get_column_mapping(mapping: List[Union[Dict[str, str], str]]) -> List[Dict[s
             else {"column": item, "key": item} for item in mapping] if mapping else []
 
 
-def generate_upsert_stmt(table: Table, business_keys: List[Dict[str, Any]], db_type: str) -> Any:
+def generate_upsert_stmt(table: Table, business_key_columns: List[str], columns: List[str], db_type: str) -> Any:
     if db_type.lower() == "postgresql":
-        for field in business_keys:
-            if not field["column"] in table.columns:
-                raise ValueError(f"{field['column']} column does not exist in {table.fullname} table")
-
-        insert_stmt = insert(table)
-
-        do_update_stmt = insert_stmt .on_conflict_do_update(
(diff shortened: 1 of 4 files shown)


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -91,15 +84,21 @@ def init(self, context: Optional[Context] = None):
...
+
+            for column in columns:
+                if not column in self.tbl.columns:
+                    raise ValueError(f"{column} column does not exist in {self.tbl.fullname} table")
 
-            self.delete_stmt = self.tbl.delete(
-                sa.and_(*[sa.text(f"{self.tbl.columns.get(field['column'])} = {sa.bindparam(field['key'])}")
-                          for field in business_keys]))
+            self.delete_stmt = self.tbl.delete().where(
+                sa.and_(*[(self.tbl.columns[column] == sa.bindparam(column)) for column in business_key_columns]))
```
zalmane, line 97: there is an edge case where you have a composite key and one of the fields is null.
it then requires:
key=incoming.key or key is null and incoming.key is null. I can open an issue for this. not urgent since we are relying on "on conflict" which requires not null constraints anyways.

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -140,7 +139,7 @@ def execute_upsert(self, records: List[Dict[str, Any]]):
             # map the record to upsert based on the mapping definitions
             # add nulls for missing mapped fields
             record_to_upsert = {}
-            for item in self.mapping:
+            for item in self.keys + self.mapping:
```
zalmane, line 142: you need to handle duplication case here where the column in both in mapping and in keys

zalmane: or decide that in such a case you just take the one from mapping (and just add that to the block's doc)

spicy-sauce: opened #214 for this.

zalmane (approved):

PR #213, opened 2022-12-17
Get_complete_job_schema

zalmane:


zalmane:
lets make sure we don't need the relpath here. I am actually not so sure why we needed it in the first place any more

spicy-sauce (approved):

PR #215, opened 2022-12-18
`write_cassandra` block #124

spicy-sauce:


core/src/datayoga_core/write_utils.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import Any, Callable, Dict, List, Union
+
+from datayoga_core.block import Block
+from datayoga_core.opcode import OpCode
+from datayoga_core.result import Result, Status
+
+logger = logging.getLogger("dy")
+
+
+def write_records(data: List[Dict[str, Any]],
```
zalmane: I would prefer not to send Callables as method parameters. It can get very messy very quickly.
Why not have this method just do the batching for you, and return 3 batches. The block can then decide what to do with it. We can then implement the deduplication here too.

core/src/datayoga_core/write_utils.py
```diff
@@ -0,0 +1,92 @@
...
+            execute_delete(records)
+        else:
+            for record in records:
+                record[Block.RESULT_FIELD] = Result(Status.REJECTED, f"{opcode} - unsupported opcode")
+            logger.warning(f"{opcode} - unsupported opcode")
+
+    return data
+
+
+def get_source_column(item: Union[Dict[str, str], str]) -> str:
```
zalmane: what do these methods do? Do we really need them as a utility method? 
in any case, I would have clear comments since it is hard to follow what is their purpose

core/src/datayoga_core/write_utils.py
```diff
@@ -0,0 +1,92 @@
...
+def get_target_column(item: Union[Dict[str, str], str]) -> str:
+    return next(iter(item.keys())) if isinstance(item, dict) else item
+
+
+def get_column_mapping(mapping: List[Union[Dict[str, str], str]]) -> List[Dict[str, str]]:
+    return [{"column": next(iter(item.keys())), "key": next(iter(item.values()))} if isinstance(item, dict)
+            else {"column": item, "key": item} for item in mapping] if mapping else []
+
+
+def get_key_values(record: Dict[str, Any], keys: List[Union[Dict[str, Any], str]]) -> Dict[str, Any]:
```
zalmane: please add clear comment explaining the purpose

core/src/datayoga_core/write_utils.py
```diff
@@ -0,0 +1,92 @@
...
+            source = get_source_column(item)
+            target = get_target_column(item)
+            record_to_upsert[target] = None if source not in record else record[source]
+
+        records_to_upsert.append(record_to_upsert)
+
+    return records_to_upsert
+
+
+def get_records_to_delete(records: List[Dict[str, Any]],
```
zalmane: aren't these a duplication of the above method that relates to upsert?

core/src/datayoga_core/blocks/cassandra/write/block.py
```diff
@@ -0,0 +1,80 @@
...
+
+        pk_clause = f"{' and '.join([column + ' = ?' for column in business_key_columns])}"
+        self.delete_stmt = f"delete from {self.keyspace}.{self.table} where {pk_clause}"
+
+        self.upsert_stmt = f"update {self.keyspace}.{self.table} set {', '.join([column + ' = ?' for column in mapping_columns])} where {pk_clause}"
+
+    async def run(self, data: List[Dict[str, Any]]) -> Tuple[List[Dict[str, Any]], List[Result]]:
+        logger.debug(f"Running {self.get_block_name()}")
+
+        data = write_utils.write_records(data, self.opcode_field, self.execute_upsert, self.execute_delete)
```
zalmane: I assume we need to open a 'transaction' here? is there such a concept in cassandra?

core/src/datayoga_core/blocks/cassandra/write/block.py
```diff
@@ -0,0 +1,80 @@
...
+        return utils.produce_data_and_results(data)
+
+    def execute_upsert(self, records: List[Dict[str, Any]]):
+        records_to_upsert = write_utils.get_records_to_upsert(records, self.keys, self.mapping)
+
+        logger.debug(f"Upserting {len(records_to_upsert)} record(s) to {self.table} table")
+        if records_to_upsert:
+            batch = BatchStatement()
```
zalmane: I have some memory saying that if the keys are distributed, writing in batches may actually slow down performance:
https://stackoverflow.com/questions/65269666/how-do-i-configure-and-execute-batchstatement-in-cassandra-correctly

core/src/datayoga_core/write_utils.py
spicy-sauce: they take an object or a string of column mapping, for example:
{ "last_name": "lname" }
or 
"last_name"

And return `last_name` for source and `lname` for target.

Since we use them in a several places in `cassandra.write` and `relational.write`, I thought it would be good to have them reusable.

core/src/datayoga_core/write_utils.py
spicy-sauce: similar, but not the same, I guess I can unify them.

core/src/datayoga_core/blocks/cassandra/write/block.py
spicy-sauce: not the same thing, see:
https://docs.datastax.com/en/cassandra-oss/2.2/cassandra/dml/dmlTransactionsDiffer.html

core/src/datayoga_core/write_utils.py
spicy-sauce: done, now I have one named `map_records`

core/src/datayoga_core/write_utils.py
spicy-sauce: removed it completely.

core/src/datayoga_core/write_utils.py
spicy-sauce: removed them completely.

core/src/datayoga_core/write_utils.py
spicy-sauce: done

core/src/datayoga_core/blocks/cassandra/write/block.py
spicy-sauce: done, replaced with preparedstatement and execute_async

core/src/datayoga_core/blocks/cassandra/write/block.py
```diff
@@ -54,27 +53,45 @@ def init(self, context: Optional[Context] = None):
...
+
+            logger.debug(f"Total {len(records)} record(s) with {opcode} opcode")
+            if opcode == OpCode.UPDATE.value:
+                self.execute_upsert(records)
+            elif opcode == OpCode.DELETE.value:
+                self.execute_delete(records)
+
         return utils.produce_data_and_results(data)
 
+    def get_future(self, stmt: PreparedStatement, record: Dict[str, Any]) -> Any:
```
zalmane, line 61: do we need a method for 2 lines if this is only called in one place? may be more maintainable to just inline

core/src/datayoga_core/write_utils.py
```diff
@@ -1,92 +1,52 @@
...
-from datayoga_core.result import Result, Status
 
 logger = logging.getLogger("dy")
 
 
-def write_records(data: List[Dict[str, Any]],
-                  opcode_field: str, execute_upsert: Callable, execute_delete: Callable) -> List[Dict[str, Any]]:
-    # group records by opcode
-    opcodes = [{"opcode": key, "records": list(result)} for key, result in groupby(
+def group_records_by_opcode(data: List[Dict[str, Any]],
```
zalmane, line 11: can we just return 3 return variables instead of a list? is it always just three? In that case, would be easier than a return type of list

core/src/datayoga_core/blocks/cassandra/write/block.py
spicy-sauce: It's called from two places.

core/src/datayoga_core/write_utils.py
spicy-sauce: done

zalmane (approved):

PR #222, opened 2022-12-21
add tests

zalmane:


core/tests/test_jmespath_expression.py
```diff
@@ -215,3 +215,8 @@ def test_jmespath_in():
 
     assert expression.search({"el": 1})
     assert not expression.search({"el": 0})
+
+def test_jmespath_nested_expression():
+    expression.compile(f"a.b")
```
spicy-sauce: "a.b" is already a string, why f-string is needed?

integration-tests/test_stdin_to_stdout.py
```diff
@@ -1,6 +1,8 @@
 import json
 import os
 
+import pytest
+
 from common.utils import run_job
```
spicy-sauce: you should install the isort extension, so the imports will be sorted the same:
```
import json
import os

import pytest
from common.utils import run_job
```

integration-tests/test_stdin_to_stdout.py
```diff
@@ -21,3 +23,25 @@ def test_stdin_to_stdout(tmpdir):
...
     os.remove(output_file)
+
+@pytest.mark.xfail(run=False)
+def test_stdin_to_stdout_filtered(tmpdir):
+    tested_data = '{"id": 121, "fname": "joe", "lname": "allen", "country_code": "US", "country_name": "united states", "gender": "M"}'
+    tested_data2 = '{"id": 122, "fname": "shmoe", "lname": "allen", "country_code": "US", "country_name": "united states", "gender": "M"}'
+    tested_data3 = '{"id": 123, "fname": "joe", "lname": "allen", "country_code": "US", "country_name": "united states", "gender": "M"}'
+
+    output_file = tmpdir.join("test_stdin_to_stdout.txt")
+    run_job("tests.stdin_to_stdout", f"echo '{tested_data}\n{tested_data2}\n{tested_data3}'", output_file)
```
spicy-sauce, line 32: why f-string is needed? it's already a string...

integration-tests/test_stdin_to_stdout.py
```diff
@@ -21,3 +23,25 @@ def test_stdin_to_stdout(tmpdir):
     assert result.get("gender") == "M"
 
     os.remove(output_file)
+
+@pytest.mark.xfail(run=False)
```
spicy-sauce: why is it marked with xfail?

spicy-sauce (changes requested):

core/tests/test_jmespath_expression.py
zalmane: fixed

integration-tests/test_stdin_to_stdout.py
zalmane: fixed

integration-tests/test_stdin_to_stdout.py
zalmane: have some conflict with this plugin. aren't they sorted now?

integration-tests/test_stdin_to_stdout.py
zalmane: it has interpolation of tested data

integration-tests/test_stdin_to_stdout.py
zalmane: removed

zalmane: it is related to the filter bug. will be resolved with that PR.

integration-tests/test_stdin_to_stdout.py
spicy-sauce: no, and `pytest` is unused. I'm fixing.

integration-tests/test_stdin_to_stdout.py
spicy-sauce: yes, my mistake.

spicy-sauce (approved):

PR #225, opened 2022-12-26
`relational.write`: mysql support

spicy-sauce:


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -7,17 +8,38 @@
...
 
 
+@unique
+class DbType(Enum):
+    MYSQL = "mysql"
+    PSQL = "postgresql"
+
+
+def get_driver_name(db_type: str) -> str:
+    if db_type == DbType.MYSQL.value:
```
zalmane, line 23: can we turn this into a dict and just pull from there? we will have tens of them

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -28,9 +50,9 @@ def init(self, context: Optional[Context] = None):
         logger.debug(f"Initializing {self.get_block_name()}")
 
         connection = utils.get_connection_details(self.properties.get("connection"), context)
-        db_type = connection.get("type")
+        db_type = connection.get("type").lower()
         engine_url = sa.engine.URL.create(
-            drivername=db_type,
+            drivername=get_driver_name(db_type),
```
zalmane, line 55: this should just be a default. we should support passing the driver name explicitly as well

core/src/datayoga_core/resources/schemas/connections.schema.json
```diff
@@ -12,7 +12,7 @@
             "type": {
               "description": "DB type",
               "type": "string",
-              "pattern": "^(?!redis|cassandra$)"
+              "enum": ["mysql", "postgresql"]
```
zalmane, line 15: what about cassandra and redis?

zalmane (changes requested):

core/src/datayoga_core/resources/schemas/connections.schema.json
spicy-sauce: they have their own section in the connections, here we used to consider everything that it's not redis and cassandra as `relational.write`, I decided to list the databases we actually support.

zalmane (approved):

PR #246, opened 2023-01-07
massive performance improvement - only bind the parts of the json that are referenced in the expression.

zalmane:


zalmane:
performance of SQL nested went from 30K/s to 600K/s

core/tests/tmp_test.py
```diff
@@ -0,0 +1,20 @@
...
+    time.sleep(int(the_time))
+    print(" Process %s\tDONE" % inputs)
+
+def mp_handler():
+    multiprocessing.set_start_method('spawn')
+    p = multiprocessing.Pool(2)
+    p.map(mp_worker, data)
+
+if __name__ == '__main__':
+    mp_handler()
```
spicy-sauce: seems that was committed by mistake, please remove.

spicy-sauce (changes requested):

spicy-sauce (approved):

PR #255, opened 2023-02-02
bug fixes

zalmane:


core/src/datayoga_core/expression.py
```diff
@@ -2,6 +2,7 @@
 import logging
 
 import sqlglot
+from sqlglot.errors import ParseError
```
spicy-sauce: not in use

core/src/datayoga_core/expression.py
```diff
@@ -14,6 +15,7 @@
 from typing import Any, Dict, List, Tuple, Union
 
 import jmespath
+
```
spicy-sauce: not needed, use the same extension we use in the project.

core/src/datayoga_core/expression.py
```diff
@@ -111,9 +115,13 @@ def compile(self, expression: str):
...
         self._column_names = set()
         for _exp in self._fields.values():
-            self._column_names.update(
-                [tuple(column.sql().replace('"', "").split("."))
-                 for column in sqlglot.parse_one("SELECT " + _exp.replace('`', '"')).find_all(sqlglot.exp.Column)])
+            try:
+                self._column_names.update(
+                    [tuple(column.sql().replace('"', "").split("."))
+                     for column in sqlglot.parse_one("SELECT " + _exp.replace('`', '"')).find_all(sqlglot.exp.Column)])
+            except Exception as pe:
```
spicy-sauce, line 120: shouldn't we catch the ParseError you imported?

spicy-sauce (changes requested):

core/src/datayoga_core/expression.py
zalmane: done

core/src/datayoga_core/expression.py
zalmane: what isn't used?

core/src/datayoga_core/expression.py
zalmane: no, since it sometimes throws runtimeerror and others

core/src/datayoga_core/expression.py
spicy-sauce: empty line, I removed it and pushed.

spicy-sauce (approved):

PR #257, opened 2023-02-02
256 upgrade cx oracle to python oracledb

vd2org:


core/src/datayoga_core/__init__.py
```diff
@@ -1,6 +1,23 @@
 import logging
 from typing import Any, Dict, List, Optional
 
+try:
+    # This hack makes it possible to use the new driver with SQLAlhemy 1.4.*
```
spicy-sauce: typo: SQLAlhemy --> SQLAlchemy

core/src/datayoga_core/__init__.py
```diff
@@ -1,6 +1,24 @@
...
+    from oracledb.exceptions import DatabaseError
+
+    oracledb.version = "8.3.0"
+    sys.modules["cx_Oracle"] = oracledb
+    try:
+        oracledb.init_oracle_client()
+    except DatabaseError:
+        pass
+except ImportError:
+    pass
```
spicy-sauce: can't we have this snippet as a util so we won't have code duplication and use it here and in the test?

spicy-sauce (dismissed):

core/src/datayoga_core/__init__.py
spicy-sauce: fixed

core/src/datayoga_core/__init__.py
```diff
@@ -1,6 +1,24 @@
 import logging
 from typing import Any, Dict, List, Optional
 
+try:
```
zalmane: @vd2org This should belong in the write.relational block. the main datayoga.core should be agnostic to what the blocks do.

zalmane: and the snippet can be moved to the utils in relational block folder (so can be reused with the test)

core/src/datayoga_core/__init__.py
vd2org: No. This code should be executed before other imports. It's necessary.

core/src/datayoga_core/__init__.py
vd2org: The same as above.

zalmane: why? I would assume it just needs to happen before the oracle import happens the first time. which can happen in the block init. can you explain?

core/src/datayoga_core/__init__.py
spicy-sauce: I understand, but why can't be a util? but this is the same issue raise by @zalmane, so please refer it there.

zalmane (approved):

zalmane (approved):

PR #267, opened 2023-02-08
Oracle import fix

vd2org:


core/src/datayoga_core/blocks/relational/__init__.py
```diff
@@ -11,7 +11,7 @@
     sys.modules["cx_Oracle"] = oracledb
     try:
         oracledb.init_oracle_client()
-    except DatabaseError:
+    except:  # noqa
         pass
```
zalmane, line 17: can you explain what this change fixes and how?

zalmane (approved):

PR #272, opened 2023-03-01
Putting things in order

vd2org:


core/src/datayoga_core/block.py
```diff
@@ -28,14 +28,14 @@ class Block(metaclass=ABCMeta):
         properties Dict[str, Any]: Block properties
     """
 
-    def __init__(self, properties: Dict[str, Any] = {}):
+    def __init__(self, properties: Dict[str, Any] = None):
```
spicy-sauce: Shouldn't it be Optional[Dict[str, Any]] then?

core/src/datayoga_core/result.py
```diff
@@ -2,18 +2,22 @@
 from enum import Enum
 from typing import Any, Dict, List, Optional
 
-Status = Enum("Status", "SUCCESS REJECTED FILTERED")
+
+class Status(Enum):
+    SUCCESS = "SUCCESS"
+    REJECTED = "REJECTED"
+    FILTERED = "FILTERED"
```
spicy-sauce: Shouldn't it be something like this:
```
@unique
class DbType(str, Enum):
    """
    Defines supported databases
    """
    CASSANDRA = "cassandra"
    MYSQL = "mysql"
    ORACLE = "oracle"
    POSTGRESQL = "postgresql"
    SQLSERVER = "sqlserver"
```

So you can access to a value without .value?

core/src/datayoga_core/step.py
```diff
@@ -11,9 +11,9 @@
 logger = logging.getLogger("dy")
 
 
-class Step():
-    def __init__(self, id: str, block: Block, concurrency=1):
-        self.id = id
+class Step:
+    def __init__(self, sid: str, block: Block, concurrency=1):
```
spicy-sauce: maybe step_id instead?

core/src/datayoga_core/step.py
```diff
@@ -30,12 +30,12 @@ async def start_pool(self):
         # start pool of workers for parallelization
         logger.debug("starting pool")
         self.queue = asyncio.Queue(maxsize=1)
-        for id in range(self.concurrency):
-            worker = self.workers[id]
+        for cid in range(self.concurrency):
```
spicy-sauce: maybe worker_id or concurrency_id instead?

core/src/datayoga_core/step_buffer.py
```diff
@@ -6,8 +6,8 @@
 logger = logging.getLogger("dy")
 
 class StepBuffer(Step):
-    def __init__(self, id, min_buffer_size=4, max_buffer_size=4, flush_ms=1000):
-        super().__init__(id, None)
+    def __init__(self, sid, min_buffer_size=4, max_buffer_size=4, flush_ms=1000):
+        super().__init__(sid, None)
```
spicy-sauce: maybe step_id instead?

core/src/datayoga_core/write_utils.py
```diff
@@ -42,9 +41,10 @@ def get_column_mapping(mapping: List[Union[Dict[str, str], str]]) -> List[Dict[s
 
 def map_record(record: Dict[str, Any],
                keys: List[Union[Dict[str, str], str]],
-               mapping: List[Union[Dict[str, str], str]] = []):
+               mapping: List[Union[Dict[str, str], str]] = None):
```
spicy-sauce: Optional[List[Union[Dict[str, str], str]]]

spicy-sauce (changes requested):

core/src/datayoga_core/block.py
vd2org: ✅

core/src/datayoga_core/result.py
vd2org: ✅

core/src/datayoga_core/step.py
vd2org: ✅

core/src/datayoga_core/step.py
vd2org: ✅

core/src/datayoga_core/step_buffer.py
vd2org: ✅

core/src/datayoga_core/write_utils.py
vd2org: ✅

core/src/datayoga_core/block.py
```diff
@@ -28,14 +30,14 @@ class Block(metaclass=ABCMeta):
         properties Dict[str, Any]: Block properties
     """
 
-    def __init__(self, properties: Dict[str, Any] = {}):
+    def __init__(self, properties: Optional[Dict[str, Any]] = None):
         """
         Constructs a block
 
         Args:
             properties (Dict[str, Any]): Block [properties]
```
spicy-sauce: properties (Optional[Dict[str, Any]]): Block [properties]

core/src/datayoga_core/step_buffer.py
```diff
@@ -6,8 +6,8 @@
 logger = logging.getLogger("dy")
 
 class StepBuffer(Step):
-    def __init__(self, id, min_buffer_size=4, max_buffer_size=4, flush_ms=1000):
-        super().__init__(id, None)
+    def __init__(self, step_id, min_buffer_size=4, max_buffer_size=4, flush_ms=1000):
```
spicy-sauce: add type for step_id

core/src/datayoga_core/step.py
```diff
@@ -11,14 +12,14 @@
 logger = logging.getLogger("dy")
 
 
-class Step():
-    def __init__(self, id: str, block: Block, concurrency=1):
-        self.id = id
+class Step:
+    def __init__(self, step_id: str, block: Block, concurrency=1):
+        self.id = step_id
```
spicy-sauce, line 17: shouldn't it be step_id as well?

core/src/datayoga_core/step_buffer.py
```diff
@@ -6,8 +6,8 @@
...
+        super().__init__(step_id, None)
```
spicy-sauce, line 11: If None is supported for block, this should be changed to in step:
```
    def __init__(self, step_id: str, block: Optional[Block], concurrency=1):
```

core/src/datayoga_core/block.py
```diff
@@ -6,13 +6,15 @@
...
 from os import path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Generator, List, Optional, TypedDict
 
 from datayoga_core import utils
 from datayoga_core.context import Context
 from datayoga_core.result import BlockResult
 from jsonschema import validate
 
+Message = TypedDict("Message", dict(msg_id=str, value=Dict[str, Any]))
```
spicy-sauce: linter:
Consider using '{"msg_id": str, "value": Dict[str, Any]}' instead of a call to 'dict'.pylint(use-dict-literal)
Expected dict or keyword parameter as second parameterPylance

spicy-sauce (changes requested):

core/src/datayoga_core/step_buffer.py
vd2org: ✅

core/src/datayoga_core/block.py
vd2org: ✅

core/src/datayoga_core/step_buffer.py
vd2org: ✅

core/src/datayoga_core/step.py
vd2org: Not necessary cuz it's member of the object.

core/src/datayoga_core/block.py
vd2org: ✅

spicy-sauce (approved):

core/src/datayoga_core/block.py
```diff
@@ -93,6 +95,22 @@ def stop(self):
     def get_block_name(self):
         return os.path.basename(os.path.dirname(sys.modules[self.__module__].__file__))
 
+    def produce(self) -> Generator[Message, None, None]:
```
zalmane: should we move to Producer altogether and separate from Block?

core/src/datayoga_core/job.py
```diff
@@ -118,11 +118,9 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
             result.processed = [Result(Status.SUCCESS, payload=row) for row in transformed_data]
         finally:
             # close the event loop
-            try:
-                loop.close()
-            except NotImplementedError as noe:
+            with suppress(NotImplementedError):
                 # for pyodide. doesn't implement loop.close, ignore
-                pass
+                loop.close()
```
zalmane, line 124: see comment above. this will break pyodide I think

core/src/datayoga_core/block.py
```diff
@@ -93,6 +95,22 @@ def stop(self):
...
+        """Produces data
+
+        Returns:
+            Generator[Message]: Produced data
+        """
+        raise NotImplementedError
+
+    def ack(self, msg_ids: List[str]):
```
zalmane: these don't belong in Block. Producer currently inherits from Block and adds those.

core/src/datayoga_core/block.py
vd2org: Probably we should split them completely. May be next time ?

core/src/datayoga_core/job.py
vd2org: Yep. I just replaced `try`/`except` clauses to more suitable `suppress`. The logic remains the same.

core/src/datayoga_core/block.py
vd2org: Yep. But it's breaks typing. It's because I move it. Will be completely solved when we split Block and Producer.

zalmane: We can't put in code that doesn't belong here because linter or typing doesn't agree. that doesn't make sense.

core/src/datayoga_core/block.py
```diff
@@ -1,119 +1,30 @@
...
+        raise NotImplementedError
 
     def stop(self):
         """
         Cleans the block connections and state
         """
         pass
-
-    def get_block_name(self):
-        return os.path.basename(os.path.dirname(sys.modules[self.__module__].__file__))
```
zalmane: why has this been removed?

core/src/datayoga_core/blocks/add_field/block.py
```diff
@@ -10,7 +10,7 @@
 logger = logging.getLogger("dy")
 
 
-class Block(DyBlock, metaclass=ABCMeta):
+class Block(DyBlock):
 
     def init(self, context: Optional[Context] = None):
         logger.debug(f"Initializing {self.get_block_name()}")
```
zalmane, line 16: this will now break since the method was removed?

core/src/datayoga_core/block.py
```diff
@@ -1,119 +1,30 @@
...
-    def ack(self, msg_ids: List[str]):
-        """Sends acknowledge for the message IDs of the records that have been processed
-
-        Args:
-            msg_ids (List[str]): Message IDs
-        """
-        raise NotImplementedError
-
-    @staticmethod
-    def create(block_name: str, properties: Dict[str, Any]) -> Block:
```
zalmane: why has this been removed?

core/src/datayoga_core/entity.py
```diff
@@ -0,0 +1,86 @@
...
+    RESULT_FIELD = f"{INTERNAL_FIELD_PREFIX}result"
+    OPCODE_FIELD = f"{INTERNAL_FIELD_PREFIX}opcode"
+    """
+    Block
+
+    Attributes:
+        properties Dict[str, Any]: Block properties
+    """
+
+    def __init__(self, properties: Optional[Dict[str, Any]] = None):
```
zalmane: they can't inherit from the same base class since Producer is about to become async. I suggest just putting the few shared lines of code into a module

vd2org: Reverted.

zalmane (approved):

PR #274, opened 2023-03-07
273 Support for reading/writing parquet files

vd2org:


docs/expressions.md
```diff
@@ -22,14 +22,15 @@ DataYoga supports both SQL and [JMESPath](https://jmespath.org/) expressions. JM
...
 | `mid`                | Returns a specified number of characters from the middle of a given text string                                                     | Input: `{"greeting": "hello world!"}` <br /> Expression: `mid(greeting, `4`, `3`)` <br /> Output: `o w`                                           |                                                                                                                                                                  |
+| `parse_json`         | Generates a random UUID4 and returns it as a string in standard format                                                              | Input: `{"data": '{"greeting": "hello world!"}'}` <br /> Expression: `parse_json(data)` <br /> Output: `{"greeting": "hello world!"}`             |                                                                                                                                                                  |
```
spicy-sauce: wrong description of this custom function

core/src/datayoga_core/jmespath_custom_functions.py
```diff
@@ -142,3 +142,10 @@ def _func_in(self, element: Any, iterable: Iterable) -> bool:
         Returns True if the iterable contains the given element.
         """
         return element in iterable
+
+    @functions.signature({"types": ["string"]})
+    def _func_json_parse(self, data: str) -> Any:
+        """\
```
spicy-sauce: Either remove \ or put the comment in one line.

core/pyproject.toml
```diff
@@ -42,6 +42,7 @@ PyMySQL = { version="^1.0.2", optional = true }
 redis = { version = "^4.3.5", optional = true }
 SQLAlchemy = { version = "^2.0.4", optional = true }
 sqlglot = "^10.4.3"
+fastparquet = "^2023.2.0"
```
spicy-sauce: let's keep it alphabetically sorted.

core/src/datayoga_core/blocks/parquet/read/block.py
```diff
@@ -0,0 +1,36 @@
...
+    def produce(self) -> Generator[Message, None, None]:
+        logger.debug("Reading parquet")
+
+        pf = ParquetFile(self.file)
+
+        count = 0
+        for df in pf.iter_row_groups():
+            for _, data in df.iterrows():
+                yield {self.MSG_ID_FIELD: str(count), **data.to_dict()}
+                count += 1
```
zalmane, line 36: what is the purpose of `count`?

core/src/datayoga_core/blocks/parquet/read/block.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def produce(self) -> Generator[Message, None, None]:
+        logger.debug("Reading parquet")
+
+        pf = ParquetFile(self.file)
+
+        count = 0
+        for df in pf.iter_row_groups():
+            for _, data in df.iterrows():
+                yield {self.MSG_ID_FIELD: str(count), **data.to_dict()}
```
zalmane, line 35: we should yield batches, not single rows

core/src/datayoga_core/blocks/parquet/read/block.py
```diff
@@ -0,0 +1,36 @@
...
+        logger.debug(f"file: {self.file}")
+
+    def produce(self) -> Generator[Message, None, None]:
+        logger.debug("Reading parquet")
+
+        pf = ParquetFile(self.file)
+
+        count = 0
+        for df in pf.iter_row_groups():
+            for _, data in df.iterrows():
```
zalmane, line 34: you can use enumerate instead of a `count` variable

core/src/datayoga_core/blocks/parquet/read/block.py
```diff
@@ -0,0 +1,36 @@
...
+
+        logger.debug(f"file: {self.file}")
+
+    def produce(self) -> Generator[Message, None, None]:
+        logger.debug("Reading parquet")
+
+        pf = ParquetFile(self.file)
+
+        count = 0
+        for df in pf.iter_row_groups():
```
spicy-sauce, line 33: I think the batch_size should be configurable.

core/src/datayoga_core/blocks/parquet/write/block.py
```diff
@@ -0,0 +1,40 @@
...
+        else:
+            self.file = os.path.join(context.properties.get("data_path"), parquet_file)
+
+        logger.debug(f"file: {self.file}")
+
+    async def run(self, data: List[Dict[str, Any]]) -> BlockResult:
+        logger.debug("Writing parquet")
+
+        # remove the internal $$msg_id column
+        out = [dict(data={i: record[i] for i in record if i != Block.MSG_ID_FIELD}) for record in data]
```
spicy-sauce: I think we should have a utility for this, as we use it in other places.

dev/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM python:3.10
```
spicy-sauce: shouldn't we use python 3.8 as well?

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/parquet/write/block.py
spicy-sauce: Consider using '{"data": {i: record[i] for i in record if i != Block.MSG_ID_FIELD}, ... }' instead of a call to 'dict'.pylint(use-dict-literal)

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/map/block.py
```diff
@@ -1,37 +1,52 @@
...
+from datayoga_core.expression import Language
 from datayoga_core.result import BlockResult, Result, Status
 
 logger = logging.getLogger("dy")
 
 
 class Block(DyBlock, metaclass=ABCMeta):
 
+    @staticmethod
+    def _prepare_expression(language: Language, expr: Union[dict, str]) -> str:
```
zalmane, line 18: please add a clear comment explaining this block and its purpose

core/src/datayoga_core/blocks/map/tests/test_map.py
```diff
@@ -51,6 +51,26 @@ async def test_map_field_double_nested_expression_jmespath():
     )
 
 
+@pytest.mark.asyncio
+async def test_map_field_nested_expression_jmespath_object():
+    block = Block(properties={"language": "jmespath",
+                              "expression": {"new_field": "`hello`", "name" : "details.fname"}})
+    block.init()
+    assert await block.run([{"details": {"fname": "john", "lname": "doe"}}]) == utils.all_success([
```
zalmane, line 59: this failed previously? I think the main thing you fixed was conversion from yaml to json that cause map to choke, no? perhaps we should add that test?

core/src/datayoga_core/blocks/parquet/write/block.py
zalmane: we should really get rid of this $$msg_id column altogether at some point. and use the Message struct where the payload is not mixed with the metadata

core/poetry.lock
```diff
@@ -168,100 +165,87 @@ pycparser = "*"
...
 name = "charset-normalizer"
-version = "3.0.1"
+version = "3.1.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = true
-python-versions = "*"
+python-versions = ">=3.7.0"
 files = [
-    {file = "charset-normalizer-3.0.1.tar.gz", hash = "sha256:ebea339af930f8ca5d7a699b921106c6e29c617fe9606fa7baa043c1cdae326f"},
```
zalmane: are all of these indeed related? or did we now pull in tons of upgraded dependencies? this would increase risk in RDI

core/pyproject.toml
zalmane: is it an optional extra? we should figure out if we want to name the extras after the block name. would make it easier to optionally install based on used blocks in the future.

core/src/datayoga_core/blocks/parquet/write/block.py
spicy-sauce: yes, let's open another issue for this - but for now, as said, it would be better to have such a utility rather code duplication.

core/pyproject.toml
```diff
@@ -50,6 +51,7 @@ oracle = ["oracledb", "SQLAlchemy"]
 pg = ["psycopg2-binary","SQLAlchemy"]
 redis = ["redis"]
 sqlserver = ["pymssql", "SQLAlchemy"]
+parquet = ["fastparquet"]
```
zalmane: should we change to `write.parquet` or `block-write.parquet`? the db drivers are actually a sub-dependency of those even, so perhaps we settle on a naming convention? If we want to start introducing many blocks, these will grow quickly and will be hard to manage. In fact, we should probably be placing them in a block descriptor json near the block at some point and compiling them here. But the first step can be a naming convention.
Thoughts?

dev/Dockerfile
vd2org: ✅

core/pyproject.toml
vd2org: ✅

core/src/datayoga_core/blocks/parquet/write/block.py
vd2org: ✅

core/src/datayoga_core/blocks/map/tests/test_map.py
vd2org: Exactly. It didn't work before.

Right. I created own converter from the json comes from yaml to the jmespath expression.

core/poetry.lock
vd2org: Yep. It just after upgrading to 3.8 and `poetry add fastparquet@2023.2.0`.

core/src/datayoga_core/blocks/map/block.py
vd2org: ✅

core/src/datayoga_core/blocks/parquet/read/block.py
vd2org: It's MSG_ID_FIELD.

core/src/datayoga_core/blocks/parquet/read/block.py
vd2org: Producer described as `def produce(self) -> Generator[Message, None, None]:`. 

Also. It called like this:

```python
        for record in self.producer.produce():
            logger.debug(f"Retrieved record:\n\t{record}")
            await self.root.process([record])
```


Should we refactor it now?

core/src/datayoga_core/blocks/parquet/read/block.py
vd2org: Nope. It's "for in for".

docs/expressions.md
vd2org: ✅

core/src/datayoga_core/jmespath_custom_functions.py
vd2org: ✅

core/src/datayoga_core/blocks/parquet/read/block.py
vd2org: `fastparquet` does not have this capability. It can read only by row_group witch is actually some kind of bucket.

vd2org: https://fastparquet.readthedocs.io/en/latest/api.html

core/src/datayoga_core/blocks/map/tests/test_map.py
zalmane: so this test should be a yaml? we need a test that simulates whatever failed previously

core/src/datayoga_core/blocks/parquet/read/block.py
zalmane: yes

zalmane: actually, let's open a separate PR

core/src/datayoga_core/blocks/parquet/write/block.py
spicy-sauce: For now:
In std/write/block.py:
```
# remove the internal $$msg_id column
filtered_record = {i: record[i] for i in record if i != Block.MSG_ID_FIELD}
```
we should have a utility for this with the code you changed in parquet/writ/block.py

core/src/datayoga_core/blocks/map/tests/test_map.py
vd2org: Not actually. Yaml parsing as object anyway as well as json. So, it's exactly the same.

core/src/datayoga_core/blocks/parquet/write/block.py
vd2org: ✅

spicy-sauce (approved):

PR #283, opened 2023-03-29
Introduce `foreach` in `relational.write` for exploding nested data #282

### LINKED ISSUE #282: Introduce `foreach` in `relational.write` for exploding nested data
Currently, the `relational.write` function writes a single record to the database, even if the input data contains nested values. This makes it difficult to store and query nested data in a relational database.

To address this, I suggest introducing a `foreach` loop in the `relational.write` function, so that it can write multiple records to the database for each nested value. This would allow for more efficient storage and querying of nested data in a relational database.

Implementation details:
1. Add a `foreach` loop to the `relational.write` function that iterates over the nested values and creates a record for each item.
2. The mapping part should contain the nested fields so those will be written as separate records to the desired DB table.
3. Update the function documentation to reflect the new behavior.

Expected Outcome:
By introducing `foreach` in `relational.write`, users will be able to store and query nested data more efficiently in a relational database. This feature will enhance the functionality and usability of the `relational.write` function.

spicy-sauce:


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -69,8 +69,9 @@ async def run(self, data: List[Dict[str, Any]]) -> BlockResult:
                     Result(status=Status.REJECTED, payload=record, message=f"unknown opcode '{opcode}'")
                     for record in opcode_groups[opcode]
                 ])
-            self.execute_upsert(opcode_groups[OpCode.CREATE] + opcode_groups[OpCode.UPDATE])
-            self.execute_delete(opcode_groups[OpCode.DELETE])
+
+            self.execute_upsert(self.explode_if_specified(opcode_groups[OpCode.CREATE] + opcode_groups[OpCode.UPDATE]))
```
zalmane: why not do the explode in the beginning before the grouping?

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -81,6 +82,18 @@ async def run(self, data: List[Dict[str, Any]]) -> BlockResult:
...
+        """Explodes the input records into multiple records if the `foreach` flag is set to true,
+            and returns the original records as is otherwise.
+
+        Args:
+            records (List[Dict[str, Any]]): The input records to process
+
+        Returns:
+            List[Dict[str, Any]]: The processed records
+        """
+        return utils.explode_records(records, self.foreach) if self.foreach else records
```
zalmane: I would do the "if" outside of the method and only call it if needed. seems strange to have an "if_specified" method

core/src/datayoga_core/utils.py
```diff
@@ -86,8 +88,8 @@ def unescape_field(field: str) -> str:
 
 
 def get_connection_details(connection_name: str, context: Context) -> Dict[str, Any]:
-    if context:
-        connection = context.properties.get("connections").get(connection_name)
+    if context and context.properties:
```
zalmane, line 91: what does this change do?

core/src/datayoga_core/utils.py
```diff
@@ -108,3 +110,23 @@ def add_uid(record: Dict[str, Any]) -> Dict[str, Any]:
 
 def remove_msg_id(record: dict) -> dict:
     return {k: v for k, v in record.items() if k != Block.MSG_ID_FIELD}
+
+
+def explode_records(records: List[Dict[str, Any]], field_expression: str) -> List[Dict[str, Any]]:
+    field_name, expression = map(str.strip, field_expression.split(":", maxsplit=1))
```
zalmane, line 127: add comments

core/src/datayoga_core/utils.py
```diff
@@ -108,3 +110,23 @@ def add_uid(record: Dict[str, Any]) -> Dict[str, Any]:
...
+    jmespath_expr = JMESPathExpression()
+    jmespath_expr.compile(expression)
+
+    exploded_records = []
+
+    for record in records:
+        exploded_fields = jmespath_expr.search(record)
+
+        if exploded_fields:
+            for exploded_field in exploded_fields:
```
zalmane: Consider changing the name - exploded_fields implies it is multiple fields, but really this holds the array of one field in multiple records

core/src/datayoga_core/write_utils.py
```diff
@@ -51,6 +52,14 @@ def map_record(record: Dict[str, Any],
 
         # columns with spaces will be later used with underscores in the bind variables
         target = (next(iter(item.keys())) if isinstance(item, dict) else item).replace(" ", "_")
-        mapped_record[target] = None if source not in record else record[source]
+
+        source_path = utils.split_field(source)
```
zalmane, line 56: what does this change do?

core/src/datayoga_core/blocks/relational/write/block.py
spicy-sauce: As explained, we first check that the opcode field exists and valid and if not reject such records. So it's inefficient to explode the records and then to reject those while we can do it in advance.

core/src/datayoga_core/blocks/relational/write/block.py
spicy-sauce: agreed, fixed.

core/src/datayoga_core/utils.py
spicy-sauce: It makes sure we won't get an exception if context.properties is None while context isn't.

core/src/datayoga_core/utils.py
spicy-sauce: done

core/src/datayoga_core/utils.py
spicy-sauce: right, changed.

core/src/datayoga_core/write_utils.py
spicy-sauce: when mapping/keys contain . in it, like address.id, it parses it as a nested field and takes the value from there instead of looking for "address.id" property, which will obviously fail.

core/src/datayoga_core/utils.py
zalmane: how is it related to this PR?

spicy-sauce: saw it and since it's a tiny fix, I decided to fix it.

zalmane (approved):

PR #288, opened 2023-04-05
csv block: batch support

vd2org:


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -59,6 +59,25 @@ def init(self, context: Optional[Context] = None):
 
     async def run(self, data: List[Dict[str, Any]]) -> BlockResult:
         logger.debug(f"Running {self.get_block_name()}")
+
+        try:
+            # try to process all records together
+            return await self._run(data)
+        except Exception as e:
```
zalmane: we need to make sure that things like retry connection are still handled the same way. Not sure how this will work if we catch all exceptions? how is this handled today?

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -59,6 +59,25 @@ def init(self, context: Optional[Context] = None):
...
+            if len(data) == 1:
+                return BlockResult(rejected=[Result(Status.REJECTED, payload=data[0], message=str(e))])
```
spicy-sauce: let's use f-string instead of str

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -59,6 +59,25 @@ def init(self, context: Optional[Context] = None):
...
+
+            # if there's an error, try to run each record separately
+            block_result = BlockResult()
+            for record in data:
+                try:
+                    block_result.extend(await self._run([record]))
+                except Exception as e:
+                    block_result.rejected.append(Result(Status.REJECTED, payload=record, message=str(e)))
```
spicy-sauce: let's use f-string instead of str

core/src/datayoga_core/result.py
```diff
@@ -22,6 +23,11 @@ class BlockResult:
     filtered: List[Result] = field(default_factory=list)
     rejected: List[Result] = field(default_factory=list)
 
+    def extend(self, other):
```
spicy-sauce: add type to other

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -59,6 +59,25 @@ def init(self, context: Optional[Context] = None):
 
     async def run(self, data: List[Dict[str, Any]]) -> BlockResult:
         logger.debug(f"Running {self.get_block_name()}")
+
+        try:
+            # try to process all records together
+            return await self._run(data)
+        except Exception as e:
+            if len(data) == 1:
```
spicy-sauce: why == 1? what if we have more records?

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/result.py
vd2org: ✅

core/src/datayoga_core/blocks/relational/write/block.py
vd2org: If we have only one record in batch there is no need to retry it, we could throw exception immediately.

core/src/datayoga_core/blocks/relational/write/block.py

Added special behavior to handle ConnectionError exception.

PR #291, opened 2023-04-10
278: Produce list of messages

### LINKED ISSUE #278: Make producer able to produce batch instead of one record
Producer described as `def produce(self) -> Generator[Message, None, None]:`. 

Also. It called like this:

```python
        for record in self.producer.produce():
            logger.debug(f"Retrieved record:\n\t{record}")
            await self.root.process([record])
```

Should we refactor it now?

_Originally posted by @vd2org in https://github.com/datayoga-io/datayoga/pull/274#discussion_r1133740225_

vd2org:


core/src/datayoga_core/blocks/files/read_csv/block.py
```diff
@@ -25,11 +25,11 @@ def init(self, context: Optional[Context] = None):
...
-    def produce(self) -> Generator[Message, None, None]:
+    def produce(self) -> Generator[List[Message], None, None]:
         logger.debug("Reading CSV")
 
         with open(self.file, "r") as read_obj:
             records = list(DictReader(read_obj))
 
         for i, record in enumerate(records):
-            yield {self.MSG_ID_FIELD: str(i), **record}
+            yield [{self.MSG_ID_FIELD: str(i), **record}]
```
zalmane, line 35: this should be a list of messages. Seems it is now one entry with a list within it? That will cause a bug.

zalmane: I wonder how it passed the tests. Are we missing coverage?

core/src/datayoga_core/blocks/parquet/read/block.py
```diff
@@ -24,13 +24,13 @@ def init(self, context: Optional[Context] = None):
...
+    def produce(self) -> Generator[List[Message], None, None]:
         logger.debug("Reading parquet")
 
         pf = ParquetFile(self.file)
 
         count = 0
         for df in pf.iter_row_groups():
             for _, data in df.iterrows():
-                yield {self.MSG_ID_FIELD: str(count), **data.to_dict()}
+                yield [{self.MSG_ID_FIELD: str(count), **data.to_dict()}]
```
zalmane, line 35: I assume we now want to product batches instead of one by one, right?

core/src/datayoga_core/blocks/relational/read/block.py
```diff
@@ -32,15 +32,15 @@ def init(self, context: Optional[Context] = None):
...
+    def produce(self) -> Generator[List[Message], None, None]:
         result = self.connection.execution_options(stream_results=True).execute(self.tbl.select())
 
         while True:
             chunk = result.fetchmany(10000)
             if not chunk:
                 break
             for row in chunk:
-                yield utils.add_uid(dict(row._asdict()))
+                yield [utils.add_uid(dict(row._asdict()))]
```
zalmane, line 43: same as above, we want to produce batches now

core/src/datayoga_core/producer.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, msg_id: str, value: Dict[str, Any]):
 class Producer(Block):
 
     @abstractmethod
-    def produce(self) -> Generator[Message, None, None]:
+    def produce(self) -> Generator[List[Message], None, None]:
```
zalmane: let's make producer async while we're making these changes. we need this for things like http producer.

zalmane (changes requested):

core/src/datayoga_core/blocks/files/read_csv/block.py
vd2org: It wasn't a list. All producers yields one message at once before.

core/src/datayoga_core/producer.py
vd2org: ✅

zalmane (approved):

PR #300, opened 2023-04-16
Producer: Http receiver

### LINKED ISSUE #299: Implement http receiver/producer

vd2org:


core/pyproject.toml
```diff
@@ -23,7 +23,7 @@ cryptography = ">=39.0.1"
 jmespath = "^1.0.0"
 jsonschema = "^4.4.0"
 orjson = "^3.8.7"
-python = "^3.7"
+python = "^3.8"
```
spicy-sauce: we can't do it, as RegisGears runs on Python 3.7.

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+
+        finally:
+            with suppress(Exception):
+                await srv.stop()
+
+    def stop(self):
+        """
+        Cleans the block connections and state
+        """
+        pass
```
spicy-sauce: If it's not implemented yet, you can remove this completely, as it will inherit it from the parent.

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Block(DyProducer, metaclass=ABCMeta):
+    port: int
+
+    def init(self, context: Optional[Context] = None):
+        logger.debug(f"Initializing {self.get_block_name()}")
+        self.port = int(self.properties.get("encoding", 8080))
+
+    async def produce(self) -> AsyncGenerator[List[Message], None]:
+        q = Queue(maxsize=1000)
```
spicy-sauce: mind to rename variable to `queue`?

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+
+logger = logging.getLogger("dy")
+
+
+class Block(DyProducer, metaclass=ABCMeta):
+    port: int
+
+    def init(self, context: Optional[Context] = None):
+        logger.debug(f"Initializing {self.get_block_name()}")
+        self.port = int(self.properties.get("encoding", 8080))
```
spicy-sauce: is it encoding? seems like a mistake...

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+
+    async def produce(self) -> AsyncGenerator[List[Message], None]:
+        q = Queue(maxsize=1000)
+
+        async def handler(request: BaseRequest) -> Response:
+            try:
+                q.put_nowait(orjson.loads(await request.read()))
+                return Response(status=200)
+            except Exception:  # noqa
+                logger.exception("Got exception while parsing request:")
```
spicy-sauce, line 35: can you also log the exception itself?

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+                return Response(status=500, text="Server error.")
```
spicy-sauce: Don't they have a constant for 500 internal error?

core/src/datayoga_core/blocks/http/receiver/block.py
```diff
@@ -0,0 +1,56 @@
...
+        logger.debug(f"Initializing {self.get_block_name()}")
+        self.port = int(self.properties.get("encoding", 8080))
+
+    async def produce(self) -> AsyncGenerator[List[Message], None]:
+        q = Queue(maxsize=1000)
+
+        async def handler(request: BaseRequest) -> Response:
+            try:
+                q.put_nowait(orjson.loads(await request.read()))
+                return Response(status=200)
```
spicy-sauce: Don't they have a constant for 200 ok?

core/src/datayoga_core/blocks/http/receiver/block.py
spicy-sauce: better to take the default 8080 from the json schema itself so if we change it to 8081 we won't need to do it twice... This is relevant to other properties and their defaults.

dev/compose-mysql.yaml
spicy-sauce, line 1: how are those compose files related to this PR?

core/src/datayoga_core/blocks/http/receiver/block.py
spicy-sauce, line 1: can you add tests? maybe by mocking a http server?

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: The `exception` method is unwinding current stacktrace implicitly. Don't need to pass exception here.

dev/compose-mysql.yaml
vd2org: I improve it as soon as it needs.

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/pyproject.toml
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: It's hard to do it here, because block don't have access to the schema. Thus, it must be refactored.

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.py
spicy-sauce: OK, we probably should take care of it in a different task.

core/src/datayoga_core/blocks/http/receiver/block.schema.json
```diff
@@ -0,0 +1,24 @@
...
+    }
+  },
+  "additionalProperties": false,
+  "examples": [
+    {
+      "host": "localhost",
+      "port": 8080
+    }
+  ]
+}
```
spicy-sauce, line 24: It seems to me that the following properties are missing:
- endpoint/uri: currently it's host:port
- https: boolean
- method: GET/POST, etc.
- Credentials for basic authentication.

dev/compose-mysql.yaml
spicy-sauce: OK, but what do they have to do with this PR? Was it taken from the previous unmerged demo PR?

vd2org: Nope. It's just the result of testing something in parallel. Probably WB and DB connectivity.

core/src/datayoga_core/blocks/http/receiver/block.py
vd2org: ✅

core/src/datayoga_core/blocks/http/receiver/block.schema.json
vd2org: I don't think we have to support https nor auth here.
The same applies to the method: it doesn't matter which one is used - producer will accept everything have a body.

Why? It's really hard to implement properly all those thinks. 
I mean, It's not only https, it's also cert based auth, external auth, LDAP and so on.

Instead of reimplementing all of this on our side we just forcing users to use appropriate component(traefik, nginx, etc) on top of runner. 

So, in short, I would like to keep this as simple as possible.

spicy-sauce (approved):

PR #303, opened 2023-04-19
Reset engine on a connection error.

### LINKED ISSUE #302: Relational: force reset connection pool on error
Relational block won't restoring after lost connection to the database.

```
sqlalchemy.exc.PendingRollbackError: Can't reconnect until invalid transaction is rolled back.  Please rollback() fully before proceeding (Background on this error at: https://sqlalche.me/e/20/8s2b)
```

vd2org:


core/src/datayoga_core/blocks/relational/write/block.py
spicy-sauce, line 1: shouldn't we apply also in `relational.read` block?

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -22,45 +23,75 @@ class Block(DyBlock, metaclass=ABCMeta):
...
+            self.keys = self.properties.get("keys")
+            self.mapping = self.properties.get("mapping")
+            self.foreach = self.properties.get("foreach")
+            tbl = sa.Table(self.table, sa.MetaData(schema=self.schema), autoload_with=engine)
 
-        if self.db_type in (relational_utils.DbType.SQLSERVER, relational_utils.DbType.ORACLE):
-            # MERGE statement requires this
-            self.connection = self.connection.execution_options(autocommit=True)
+            logger.debug(f"Connecting to {self.db_type}")
+            connection = engine.connect()
```
spicy-sauce: shouldn't you try to connect first before everything done in lines 37-44?

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -22,45 +23,75 @@ class Block(DyBlock, metaclass=ABCMeta):
...
+            self.engine.dispose()
+
+        self.business_key_columns = None
+        self.mapping_columns = None
+        self.columns = None
+        self.delete_stmt = None
+        self.upsert_stmt = None
+        self.tbl = None
+        self.connection = None
+        self.engine = None
```
spicy-sauce: you can do this instead, IMO, so you don't need to specify them all and maintain it here when a new class member is added:
```
  for attr in vars(self):
        setattr(self, attr, None)
```

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/relational/write/block.py
vd2org: Sure. Haven't checked yet. We will do this in a separate PR.

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/blocks/relational/write/block.py
spicy-sauce: OK, please open such an issue, so it won't be forgotten.

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -18,49 +19,75 @@
...
+            self.keys = self.properties.get("keys")
+            self.mapping = self.properties.get("mapping")
+            self.foreach = self.properties.get("foreach")
+            tbl = sa.Table(self.table, sa.MetaData(schema=self.schema), autoload_with=engine)
 
-        if self.opcode_field:
-            self.business_key_columns = [column["column"] for column in write_utils.get_column_mapping(self.keys)]
-            self.mapping_columns = [column["column"] for column in write_utils.get_column_mapping(self.mapping)]
+            self.engine = engine
+            self.connection = connection
```
spicy-sauce: I think those should be declared after line 40 (the connection created successfully) for better readability.

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -18,49 +19,75 @@
...
-            # MERGE statement requires this
-            self.connection = self.connection.execution_options(autocommit=True)
+            self.schema = self.properties.get("schema")
+            self.table = self.properties.get("table")
+            self.opcode_field = self.properties.get("opcode_field")
+            self.load_strategy = self.properties.get("load_strategy")
+            self.keys = self.properties.get("keys")
+            self.mapping = self.properties.get("mapping")
+            self.foreach = self.properties.get("foreach")
+            tbl = sa.Table(self.table, sa.MetaData(schema=self.schema), autoload_with=engine)
```
spicy-sauce: maybe better to replace with `self.engine` to be consistent (after moving it up as suggested in the other comment).

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/relational/write/block.py
vd2org: ✅

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/blocks/relational/write/block.py

spicy-sauce (approved):

PR #304, opened 2023-04-19
`relational.write`: special cases for oracle and mssql server

### LINKED ISSUE #301: `relational.write` - Special cases for Oracle and MSSql server
In Oracle, when DB is not available a `sqlalchemy.exc.DatabaseError` is thrown and not `OperationalError`.
We should catch this for Oracle only (as for other databases, this exception might be different or at least, need to check it, so you can catch them both together if this assumption is wrong).

For MSSql server `OperationalError` is thrown in some cases not related to connection.

### LINKED ISSUE #306: Performance issue with SQLA and MySQL 8.0.17+

vd2org:


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -39,6 +39,11 @@ def setup_engine(self):
             logger.debug(f"Connecting to {self.db_type}")
             self.connection = self.engine.connect()
 
+            # Disable the new MySQL 8.0.17+ default behavior of requiring an alias for ON DUPLICATE KEY UPDATE
+            # This behavior is not supported by pymysql driver
+            if self.db_type == relational_utils.DbType.MYSQL:
```
spicy-sauce: shouldn't we check it according to the driver? so if other driver than pymysql is used, the behavior might be different?

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -163,7 +173,26 @@ def execute(self, statement: Any, records: List[Dict[str, Any]]) -> CursorResult
...
+            raise
+
+    def handle_mssql_operational_error(self, e):
+        """Handling specific MSSQL cases: Conversion failed(245) and Truncated data(2628)"""
+        if e.orig.args[0] in (245, 2628):
+            raise
+
+    def hande_oracle_database_error(self, e):
+        """Handling specific OracleDB cases: Network failure(DPY-4011) and Database restart(ORA-01089)"""
+        if "DPY-4011" in str(e) or "ORA-01089" in str(e):
```
spicy-sauce: use f-string:
if "DPY-4011" in f"{e}" or "ORA-01089" in f"{e}":

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -163,7 +173,26 @@ def execute(self, statement: Any, records: List[Dict[str, Any]]) -> CursorResult
...
+
+            raise
+
+    def handle_mssql_operational_error(self, e):
+        """Handling specific MSSQL cases: Conversion failed(245) and Truncated data(2628)"""
+        if e.orig.args[0] in (245, 2628):
+            raise
+
+    def hande_oracle_database_error(self, e):
+        """Handling specific OracleDB cases: Network failure(DPY-4011) and Database restart(ORA-01089)"""
```
spicy-sauce: Handling specific OracleDB cases: Network failure (DPY-4011) and Database restart (ORA-01089)

core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -163,7 +173,26 @@ def execute(self, statement: Any, records: List[Dict[str, Any]]) -> CursorResult
...
+            self.dispose_engine()
+            raise ConnectionError(e)
+        except DatabaseError as e:
+            if self.db_type == relational_utils.DbType.ORACLE:
+                self.hande_oracle_database_error(e)
+
+            raise
+
+    def handle_mssql_operational_error(self, e):
+        """Handling specific MSSQL cases: Conversion failed(245) and Truncated data(2628)"""
```
spicy-sauce: """Handling specific MSSQL cases: Conversion failed (245) and Truncated data (2628)"""

dev/compose-mssql.yaml
```diff
@@ -17,4 +17,4 @@ services:
 
 networks:
   service:
-    external: true
\ No newline at end of file
+    external: true
```
spicy-sauce, line 20: what was changed here?

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/blocks/relational/write/block.py

core/src/datayoga_core/blocks/relational/write/block.py

dev/compose-mssql.yaml
vd2org: Prevents compose to create a new network. Use the existing one instead.

core/src/datayoga_core/blocks/relational/write/block.py

spicy-sauce (approved):

PR #313, opened 2023-06-14
Use prettier for the generated docs.

vd2org:


spicy-sauce:
No need at this point.

PR #315, opened 2023-07-12
azure event hub reader

### LINKED ISSUE #314: Azure Event Hubs Reader

spicy-sauce:


core/src/datayoga_core/blocks/azure/read_event_hub/block.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Starts the event receiving process and yield batches of messages.
+
+        Yields:
+            AsyncGenerator[List[Message], None]: A generator of message batches.
+        """
+        logger.debug(f"Running {self.get_block_name()}")
+
+        logger.debug("Starting event receiving process")
+        asyncio.create_task(self.receive_batch())
```
vd2org, line 55: Don't leave a task without a reference. The Eventloop only keeps weak references to tasks, so it might be collected by GC at anytime.

Need something like this:
`self._task = create_task(...)`

https://docs.python.org/3/library/asyncio-task.html?highlight=create_task#asyncio.create_task

vd2org (changes requested):

core/src/datayoga_core/blocks/azure/read_event_hub/block.py
spicy-sauce: good catch, thanks, done.

vd2org (approved):

PR #317, opened 2023-07-17
Implemented async version of transform function.

### LINKED ISSUE #316: Make able `Job.transform` to use external event loop

vd2org:


core/src/datayoga_core/producer.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, msg_id: str, value: Dict[str, Any]):
 class Producer(Block):
 
     @abstractmethod
-    async def produce(self) -> AsyncGenerator[List[Message], None]:
+    def produce(self) -> AsyncGenerator[List[Message], None]:
```
spicy-sauce: In this case, doesn't it mean that all the methods that implement this abstract method should have the same signature?

for example,
    async def produce(self) -> AsyncGenerator[List[Message], None]:

Method 'produce' was expected to be 'non-async', found it instead as 'async'Pylint[W0236:invalid-overridden-method](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/invalid-overridden-method.html)

Other references:
image

core/src/datayoga_core/producer.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, msg_id: str, value: Dict[str, Any]):
...
         """Produces data
 
         Returns:
```
spicy-sauce, line 19: not updated return type:
```
Generator[Message]: Produced data
```

Should be:
```
AsyncGenerator[List[Message], None]
```

Or even,
```
        Yields:
            AsyncGenerator[List[Message], None]: A generator of message batches.
```

core/src/datayoga_core/job.py
```diff
@@ -125,6 +125,49 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
 
         return result
 
+    async def async_transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> JobResult:
```
spicy-sauce: boolean --> bool

core/src/datayoga_core/job.py
```diff
@@ -125,6 +125,49 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
...
+        """
+        Transforms data(async version)
```
spicy-sauce: Add space and period:
Transforms data (async version).

core/src/datayoga_core/job.py
```diff
@@ -125,6 +125,49 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
...
+
+        Args:
+            data (List[Dict[str, Any]]): Data
```
spicy-sauce: add period.

core/src/datayoga_core/job.py
```diff
@@ -125,6 +125,49 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
...
+            deepcopy: if True, performs a deepcopy before modifying records. otherwise, modifies in place. can affect performance.
```
spicy-sauce: add type:
deepcopy (bool): ...

spicy-sauce: Change description to:
```
If set to True, performs a deepcopy before modifying records; otherwise, it modifies them in place. This can affect performance.
```

core/src/datayoga_core/job.py
spicy-sauce: can we have shared logic in `async_transform` and `transform` in a utility called from those methods rather than repeating ourselves? 95% of the logic is the same, will be tricky to maintain.

spicy-sauce: can we have test(s) to cover this new functionality?

core/src/datayoga_core/job.py
```diff
@@ -125,6 +125,49 @@ def transform(self, data: List[Dict[str, Any]], deepcopy: boolean = True) -> Job
...
+
+        Returns:
+            JobResult: Job result
```
spicy-sauce: add period.

spicy-sauce (changes requested):

core/src/datayoga_core/job.py
spicy-sauce: one way suggested by chatgpt:
```
import asyncio

def transform(self, data: List[Dict[str, Any]], deepcopy: bool = True) -> JobResult:
    """
    Transforms data

    Args:
        data (List[Dict[str, Any]]): Data
        deepcopy: if True, performs a deepcopy before modifying records. otherwise, modifies in place. can affect performance.

    Returns:
        JobResult: Job result
    """
    if not self.initialized:
        logger.debug("job has not been initialized yet, initializing...")
        self.init()
    
    # Shared logic for both synchronous and asynchronous versions
    transformed_data = marshal.loads(marshal.dumps(data)) if deepcopy else data
    result = JobResult()
    
    # create an event loop for the synchronous version
    loop = asyncio.new_event_loop()

    try:
        # Call the async_transform method
        result = loop.run_until_complete(self.async_transform(transformed_data, deepcopy))
    finally:
        # close the event loop
        with suppress(NotImplementedError):
            # for pyodide. doesn't implement loop.close, ignore
            loop.close()

    return result

async def async_transform(self, data: List[Dict[str, Any]], deepcopy: bool = True) -> JobResult:
    """
    Transforms data (async version)

    Args:
        data (List[Dict[str, Any]]): Data
        deepcopy: if True, performs a deepcopy before modifying records. otherwise, modifies in place. can affect performance.

    Returns:
        JobResult: Job result
    """
    if not self.initialized:
        logger.debug("job has not been initialized yet, initializing...")
        self.init()

    # Shared logic for both synchronous and asynchronous versions
    transformed_data = marshal.loads(marshal.dumps(data)) if deepcopy else data
    result = JobResult()

    for step in self.steps:
        try:
            if len(transformed_data) == 0:
                # in case all records have been filtered, stop sending
                break
            processed, filtered, rejected = await step.block.run(transformed_data)
            result.filtered.extend(filtered)
            result.rejected.extend(rejected)
            transformed_data = [result.payload for result in processed]
        except ConnectionError as e:
            # connection errors are thrown back to the caller to handle
            raise e
        except Exception as e:
            # other exceptions are rejected
            logger.error(f"Error while transforming data: {e}")
            result.rejected.extend(
                [Result(Status.REJECTED, payload=row, message=f"{e}") for row in transformed_data])
            return result

    # the processed records are those that make it to the end
    result.processed = [Result(Status.SUCCESS, payload=row) for row in transformed_data]

    return result

```

Not sure if we want to create another event loop in the synced version, but an idea to consider.

spicy-sauce: This one doesn't create another event loop:
```
import asyncio

def transform(self, data: List[Dict[str, Any]], deepcopy: bool = True) -> JobResult:
    """
    Transforms data

    Args:
        data (List[Dict[str, Any]]): Data
        deepcopy: if True, performs a deepcopy before modifying records. otherwise, modifies in place. can affect performance.

    Returns:
        JobResult: Job result
    """
    if not self.initialized:
        logger.debug("job has not been initialized yet, initializing...")
        self.init()

    # Shared logic for both synchronous and asynchronous versions
    transformed_data = marshal.loads(marshal.dumps(data)) if deepcopy else data
    result = asyncio.run(self.async_transform(transformed_data, deepcopy))
    
    return result

async def async_transform(self, data: List[Dict[str, Any]], deepcopy: bool = True) -> JobResult:
    """
    Transforms data (async version)

    Args:
        data (List[Dict[str, Any]]): Data
        deepcopy: if True, performs a deepcopy before modifying records. otherwise, modifies in place. can affect performance.

    Returns:
        JobResult: Job result
    """
    if not self.initialized:
        logger.debug("job has not been initialized yet, initializing...")
        self.init()

    # Shared logic for both synchronous and asynchronous versions
    transformed_data = marshal.loads(marshal.dumps(data)) if deepcopy else data
    result = JobResult()

    for step in self.steps:
        try:
            if len(transformed_data) == 0:
                # in case all records have been filtered, stop sending
                break
            processed, filtered, rejected = await step.block.run(transformed_data)
            result.filtered.extend(filtered)
            result.rejected.extend(rejected)
            transformed_data = [result.payload for result in processed]
        except ConnectionError as e:
            # connection errors are thrown back to the caller to handle
            raise e
        except Exception as e:
            # other exceptions are rejected
            logger.error(f"Error while transforming data: {e}")
            result.rejected.extend(
                [Result(Status.REJECTED, payload=row, message=f"{e}") for row in transformed_data])
            return result

    # the processed records are those that make it to the end
    result.processed = [Result(Status.SUCCESS, payload=row) for row in transformed_data]

    return result
```

See if this can fit or suggest something else to refactor.

core/src/datayoga_core/producer.py
vd2org: Reverted.

core/src/datayoga_core/producer.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: It's temporary solution. Sync version must be removed as soon as we will be ready.

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: ✅

core/src/datayoga_core/job.py
vd2org: There is no new logic. Current tests must be refactored to async as soon as we'll finish refactor.

spicy-sauce (approved):

PR #325, opened 2023-09-06
322: redis.lookup block(nested assignments)

vd2org:


core/src/datayoga_core/blocks/redis/lookup/block.py
```diff
@@ -26,7 +26,7 @@ def init(self, context: Optional[Context] = None):
             connection.get("password")
         )
 
-        self.field = self.properties.get("field")
+        self.field_path = [utils.unescape_field(f) for f in utils.split_field(self.properties.get("field"))]
```
spicy-sauce: can we replace f with field?

core/src/datayoga_core/utils.py
```diff
@@ -79,11 +79,16 @@ def get_resource_path(relative_path: str) -> str:
         return path.join(os.path.dirname(__file__), "resources", relative_path)
 
 
-def split_field(field: str) -> List[str]:
-    return re.split(r"(?<!\\)\.", field)
+def split_field(field: str, *, __expression=re.compile(r"(?<!\\)\.")) -> List[str]:
```
spicy-sauce, line 82: why do we pass the expression as an argument? It's not like we call it with other pattern...

core/src/datayoga_core/utils.py
```diff
@@ -79,11 +79,16 @@ def get_resource_path(relative_path: str) -> str:
...
+    r"""Split string by dots like str.split(), but allows escape a dot by the backslash.
```
spicy-sauce: r"""Splits a string by dots, similar to `str.split()`, but allows escaping a dot with a backslash (\).

spicy-sauce: Why r is needed here?

core/src/datayoga_core/utils.py
```diff
@@ -79,11 +79,16 @@ def get_resource_path(relative_path: str) -> str:
...
+    Example:
+        a.b\.c.d => ["a", "b\\.c", "d"]
+    """
+    return __expression.split(field)
 
 
 def unescape_field(field: str) -> str:
+    r"""Replaces "\\." by "."."""
```
spicy-sauce, line 91: why r is needed here?

spicy-sauce (changes requested):

core/src/datayoga_core/blocks/redis/lookup/block.py
vd2org: ✅

core/src/datayoga_core/utils.py
vd2org: It's the way to avoid using an external variable to store the expression. The default value acts like a cache or a global variable and double underscore indicates that this arg must not be used by a caller. Compiled expressions works faster.

core/src/datayoga_core/utils.py
vd2org: ✅

vd2org: To not escape backslashes in the docstring. Makes it more readable. Otherwise it should be escaped like for normal strings.

core/src/datayoga_core/utils.py
vd2org: To not escape backslashes in the docstring. Makes it more readable. Otherwise it should be escaped like for normal strings.

spicy-sauce (approved):

PR #326, opened 2023-09-07
Dry mode for `redis.lookup`

vd2org:


core/src/datayoga_core/blocks/redis/lookup/block.py
```diff
@@ -19,12 +19,15 @@ def init(self, context: Optional[Context] = None):
...
-        )
+        if self.properties.get("dry"):
+            self.redis_client = None
+        else:
+            self.redis_client = redis_utils.get_client(
+                connection.get("host"),
+                connection.get("port"),
+                connection.get("user"),
+                connection.get("password")
+            )
```
spicy-sauce: why not just:
```
if not self.properties.get("dry"):
            self.redis_client = redis_utils.get_client(
                connection.get("host"),
                connection.get("port"),
                connection.get("user"),
                connection.get("password")
            )
```

core/src/datayoga_core/blocks/redis/lookup/block.py
spicy-sauce, line 1: Should be added to the jsonschema of the block, no?

spicy-sauce (changes requested):

spicy-sauce (approved):

PR #339, opened 2023-11-06
`http.write` block #337

### LINKED ISSUE #337: `http.write` block

spicy-sauce:


core/src/datayoga_core/blocks/http/write/block.py
```diff
@@ -0,0 +1,125 @@
...
+                elif self.method == "POST":
+                    response = requests.post(url, data=payload, params=query_params,
+                                             headers=headers, timeout=self.timeout)
+                elif self.method == "PUT":
+                    response = requests.put(url, data=payload, params=query_params,
+                                            headers=headers, timeout=self.timeout)
+                elif self.method == "DELETE":
+                    response = requests.delete(url, params=query_params, headers=headers, timeout=self.timeout)
+
+                if response:
```
vd2org: Could the `response` be false?

vd2org (changes requested):

core/src/datayoga_core/blocks/http/write/block.py
```diff
@@ -0,0 +1,125 @@
...
+            try:
+                url = f"{self.base_uri}/{request_configs['endpoint'][i].lstrip('/')}"
+                headers = {key: value[i] for key, value in request_configs["headers"].items()}
+                query_params = {key: value[i] for key, value in request_configs["query_params"].items()}
+                payload = {key: value[i] for key, value in request_configs["payload"].items()}
+
+                logger.debug(
+                    f"Sending HTTP {self.method} request to: {url}\nheaders:{headers}\n\tquery_params: {query_params}\n\tpayload: {payload}")
+
+                if self.method == "GET":
```
vd2org: And… why not `aiohttp` or `httpx` which are async ?

vd2org (changes requested):

core/src/datayoga_core/blocks/http/write/block.py
spicy-sauce: Replaced with aiohttp, so irrelevant.

core/src/datayoga_core/blocks/http/write/block.py
spicy-sauce: Done.

vd2org (approved):

PR #374, opened 2024-09-19
`relational.write` - SQL statements are executed individually instead of in bulk for MySQL #373

### LINKED ISSUE #373: `relational.write` - SQL statements are executed individually instead of in bulk for MySQL

spicy-sauce:


spicy-sauce:
abandoning due to licensing issues of this mysql driver.

PR #375, opened 2024-09-19
`relational.write` - SQL statements are executed individually instead of in bulk for MySQL #373

spicy-sauce:


core/src/datayoga_core/blocks/relational/utils.py
```diff
@@ -77,9 +78,34 @@ def get_engine(connection_name: str, context: Context, autocommit: bool = True)
...
+def create_pymysql_dialect_configuration(engine: Engine) -> Callable[[DBAPIConnection, ConnectionPoolEntry], None]:
+    """Creates a function to configure the PyMySQL dialect.
+
+    Args:
+        engine (Engine): The SQLAlchemy engine to configure.
+
+    Returns:
+        Callable[[DBAPIConnection, ConnectionPoolEntry], None]: A function that configures PyMySQL dialect settings.
+    """
+    def configure_pymysql_dialect(dbapi_connection: DBAPIConnection, connection_record: ConnectionPoolEntry):
```
zalmane, line 96: Why do you need a submethod in there?

spicy-sauce: Just because it is tied to the `create_pymysql_dialect_configuration` function, which requires the `engine` to be passed as an argument.

PR #380, opened 2024-12-02
handle MySQL connection lost errors

### LINKED ISSUE #379: `relational.write`: Handle MySQL server lost errors (CR_SERVER_LOST, etc.) as ConnectionError

spicy-sauce:


.github/workflows/integration-tests.yml
```diff
@@ -20,6 +20,11 @@ jobs:
         with:
           python-version: "3.8"
 
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y freetds-dev libkrb5-dev
```
zalmane, line 26: relevant?

spicy-sauce: These system dependencies are relevant when `pymssql` needs to be built from source because no pre-built wheel is available for your specific Python version and platform. Without them it's currently broken, so not sure what was changed in the runner.

zalmane (approved):

PR #382, opened 2024-12-16
Issues with Redis's get_client method #381

### LINKED ISSUE #381: Issues with Redis's get_client method
Those should be added:
-  socket_timeout
-  socket_connect_timeout
-  socket_keepalive
-  health_check_interval

ssl parameters should be set like this:
```python
ssl=cacert is not None or cert is not None or key is not None,
ssl_ciphers=SSL_CIPHERS,  # Customize TLS ciphers (Python 3.10+)
```

This should be removed:
```python
ssl_cert_reqs=ssl.CERT_REQUIRED if cacert else ssl.CERT_NONE,
```

Also, new redis and hiredis should be used.

spicy-sauce:


.github/workflows/integration-tests.yml
```diff
@@ -20,10 +20,10 @@ jobs:
         with:
           python-version: "3.8"
 
-      - name: Install system dependencies
+      - name: Install system and Oracle dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y freetds-dev libkrb5-dev
+          sudo apt-get install -y freetds-dev libkrb5-dev libaio1
```
zalmane: related?

schemas/connections.schema.json
```diff
@@ -414,18 +414,50 @@
...
+              "default": true
+            },
+            "health_check_interval": {
+              "description": "Interval for health checks in seconds",
+              "type": "number",
+              "minimum": 0,
+              "default": 60
             }
           },
           "additionalProperties": false,
```
zalmane, line 442: do we have extra_args? I thought we have it for other DBs.

core/src/datayoga_core/blocks/redis/utils.py
```diff
@@ -36,23 +41,26 @@ def get_client(connection_details: Dict[str, Any]) -> Redis:
...
             ssl_password=key_password,
             ssl_certfile=cert,
-            ssl_cert_reqs=ssl.CERT_REQUIRED if cacert else ssl.CERT_NONE,
             ssl_ca_certs=cacert,
+            ssl_ciphers=SSL_CIPHERS,  # Customize TLS ciphers (Python 3.10+)
             decode_responses=True,
-            client_name="datayoga"
+            client_name="datayoga",
+            socket_timeout=connection_details.get("socket_timeout", 10.0),
+            socket_connect_timeout=connection_details.get("socket_connect_timeout", 2.0),
```
zalmane, line 58: are these the defaults we use today?

spicy-sauce: We use those in RDI. We haven't used these new properties at all up until now.

schemas/connections.schema.json
spicy-sauce: In other DBs we do, but this, since it's a Redis connection identical to the one that is used in RDI, we have those properties in the connection level.

.github/workflows/integration-tests.yml
spicy-sauce: Failed without it, but on the other hand, then I pinned ubuntu-22.04, so probably not needed, removed, let's see that the tests still pass.

schemas/connections.schema.json
zalmane: OK. We can add later if needed

zalmane (approved):

PR #391, opened 2025-01-08
`relational.write` - When a batch fails, insert the records one by one #378

### LINKED ISSUE #378: `relational.write` - When a batch fails, insert the records one by one
So only the problematic record will be rejected and not the entire batch it's part of as result.

spicy-sauce:


core/src/datayoga_core/blocks/relational/write/block.py
```diff
@@ -162,6 +166,43 @@ def generate_upsert_stmt(self) -> Any:
...
+            Tuple[List[Result], List[Result]]: Processed and rejected records.
+        """
+        processed_records: List[Result] = []
+        rejected_records: List[Result] = []
+
+        try:
+            execute_method(records)
+            processed_records.extend([Result(Status.SUCCESS, payload=record) for record in records])
+        except Exception as batch_error:
+            logger.warning(f"Batch operation failed: {batch_error}")
```
zalmane: add "will be retried one by one" or something to indicate it is a warning and not an error

spicy-sauce: done

zalmane (approved):

PR #398, opened 2025-12-26
Feature/when block in redis lookup

ZdravkoDonev-redis:
In order to reduce latency, we can avoid doing lookups altogether by looking in some of the fields in the input data and deciding whether we should doo a lookup in advance. Thus with this commit, a 'when' field is introduced to support this functionality


spicy-sauce (approved):

spicy-sauce:
Thanks a lot, @ZdravkoDonev-redis! This PR raised a few issues as this was the first one done on a forked branch.

The feature itself looks good - the when condition for redis.lookup is well implemented with:
- Clean schema definition supporting both jmespath and sql expression languages
- Proper error handling (lookup is skipped if condition evaluation fails)
- Efficient pipeline usage - only records passing the condition trigger Redis calls
- Comprehensive test coverage for both languages and edge cases

Infrastructure fixes included in this PR:

Since this was our first fork PR, we discovered several CI/workflow issues:
  1. Checkout for forks - Workflows couldn't fetch the branch from a forked repo. Fixed by adding the repository parameter to actions/checkout.
  2. Auto-push for fork PRs - GitHub doesn't expose secrets to fork PR workflows, so the auto-commit of generated schemas/docs fails. We now detect fork PRs and display a helpful error message with the exact command to run locally.
  3. Non-deterministic schema generation - The generated JSON schemas differed between local and CI runs due to Python's hash randomization and unordered file iteration. Fixed by sorting block types and using sort_keys=True in JSON output.
  4. Shared generation scripts - Created `scripts/generate-jsonschemas.sh` and `scripts/generate-docs.sh` to ensure local and CI generation produce identical results.

PR #399, opened 2026-01-03
Update orjson version CVE issue, change Python min version to 3.8

ZdravkoDonev-redis:
We had a JIT CVE reported once I wanted to use the latest datayoga version after my pervious PR:

```
Security control: Software Component Analysis Poetry

Orjson Does Not Limit Recursion For Deeply Nested Json Documents

orjson.loads in orjson before 3.9.15 does not limit recursion for deeply nested JSON documents.

This vulnerability is introduced by the direct library datayoga-core@1.132.0, which brings in the issue through its dependencies:
datayoga-core@1.132.0 -> orjson@3.8.7

Is fix available? Yes, this issue was fixed in version 3.9.15

Severity: HIGH

```

[Learn more about this issue](https://nvd.nist.gov/vuln/detail/CVE-2024-27454)

In order to udpate the orjson version, I had to also udpate the Python version.


core/pyproject.toml
```diff
@@ -22,9 +22,9 @@ cryptography = ">=39.0.1"
 Jinja2 = "^3.1.2"
 jmespath = "^1.0.0"
 jsonschema = "^4.4.0"
-orjson = "^3.8.7"
+orjson = "^3.9.15"
 prometheus-client = "^0.16.0"
-python = "^3.7"
+python = "^3.8"
```
spicy-sauce, line 27: We pinned it to 3.7 as RedisGears uses it for WriteBehind, you sure it won't break anything on your end?

ZdravkoDonev-redis: We will definitely not update the dayayoga library in the Preview and Write behind versions, so we're good.

spicy-sauce: In this case, please also apply these changes:

```
  | File                | Line | Current         | Should be       |
  |---------------------|------|-----------------|-----------------|
  | cli/pyproject.toml  | 14   | Python :: 3.7   | Python :: 3.8   |
  | cli/pyproject.toml  | 23   | python = "^3.7" | python = "^3.8" |
  | core/pyproject.toml | 14   | Python :: 3.7   | Python :: 3.8   |
```

also regenerate `poetry.lock` for cli and core.

spicy-sauce (changes requested):

core/pyproject.toml
ZdravkoDonev-redis: Done

spicy-sauce (approved):

